PR #439, opened 2023-09-20
Fix: Remove Redundant Conditional in Search Algorithm

gandalf-repo:


samwise-repo1 (approved):

PR #466, opened 2023-09-20
WEB-408 Fix: Resolve Missing Icons in Navigation Bar

gandalf-repo:


frodo-repo (changes requested):

PR #475, opened 2023-09-20
WEB-645 Refactor: Switch from Callbacks to Promises

gandalf-repo:


frodo-repo (changes requested):
Great work on the UI, it's looking very clean and user-friendly!

PR #486, opened 2023-09-20
WEB-458 Style: Introduce New Color Palette for UI

gandalf-repo:


samwise-repo1 (changes requested):

PR #499, opened 2023-09-20
WEB-672 Fix: Resolve Missing Icons in Navigation Bar

samwise-repo1:


frodo-repo (changes requested):

PR #554, opened 2023-09-20
WEB-678 Bug: Fix Broken Image Links in Frontend

gandalf-repo:


frodo-repo (changes requested):

PR #573, opened 2023-09-20
WEB-695 Bug: Fix Broken Image Links in Frontend

samwise-repo1:


gandalf-repo (changes requested):

PR #576, opened 2023-09-20
WEB-671 Fix: Resolve Missing Icons in Navigation Bar

gandalf-repo:


samwise-repo1 (changes requested):

PR #589, opened 2023-09-21
WEB-684 Feature: Introduce Mobile-First Responsive Design

frodo-repo:


samwise-repo1 (changes requested):

PR #652, opened 2023-09-21
WEB-705 Style: Introduce New Color Palette for UI

frodo-repo:


gandalf-repo (commented):

PR #720, opened 2023-09-21
WEB-741 Refactor: Switch from Callbacks to Promises

gandalf-repo:
This PR fixes incomplete error handling in the API, ensuring that all error cases are appropriately handled.


frodo-repo (approved):

PR #744, opened 2023-09-21
WEB-775 Style: Introduce New Color Palette for UI

gandalf-repo:


frodo-repo (commented):

PR #747, opened 2023-09-21
WEB-740 Style: Introduce New Color Palette for UI

frodo-repo:


gandalf-repo (commented):

PR #748, opened 2023-09-21
WEB-729 Style: Introduce New Color Palette for UI

gandalf-repo:


samwise-repo1 (approved):

PR #751, opened 2023-09-21
WEB-765 Style: Introduce New Color Palette for UI

frodo-repo:


gandalf-repo (changes requested):

PR #778, opened 2023-09-22
WEB-751 Bug: Fix Broken Image Links in Frontend

frodo-repo:


gandalf-repo (changes requested):

PR #793, opened 2023-09-22
WEB-805 Style: Introduce New Color Palette for UI

gandalf-repo:


samwise-repo1 (changes requested):

PR #802, opened 2023-09-22
WEB-810 Style: Introduce New Color Palette for UI

gandalf-repo:


samwise-repo1 (approved):

PR #806, opened 2023-09-22
WEB-776 Fix: Address Incomplete Error Handling in API

frodo-repo:


samwise-repo1 (changes requested):

PR #812, opened 2023-09-22
WEB-853 Style: Introduce New Color Palette for UI

frodo-repo:


samwise-repo1 (approved):

PR #821, opened 2023-09-22
WEB-791 Refactor: Split Monolithic Service into Microservices

frodo-repo:


gandalf-repo (approved):

PR #882, opened 2023-09-22
WEB-843 Chore: Update Project Dependencies

frodo-repo:


samwise-repo1 (approved):

PR #908, opened 2023-09-22
WEB-902 Refactor: Split Monolithic Service into Microservices

gandalf-repo:


samwise-repo1 (changes requested):
The CSS rules here seem to be redundant. Can you please review and remove unnecessary ones?

PR #997, opened 2023-09-23
WEB-957 Fix: Resolve Missing Icons in Navigation Bar

gandalf-repo:


samwise-repo1 (changes requested):

PR #1021, opened 2023-09-23
WEB-944 Fix: Resolve Missing Icons in Navigation Bar

samwise-repo1:


gandalf-repo (approved):

PR #1034, opened 2023-09-23
WEB-964 Feature: Introduce Mobile-First Responsive Design

samwise-repo1:


frodo-repo (changes requested):

PR #1035, opened 2023-09-23
WEB-967 Fix: Remove Redundant Conditional in Search Algorithm

gandalf-repo:


frodo-repo (changes requested):

PR #1045, opened 2023-09-23
WEB-943 Feature: Introduce Mobile-First Responsive Design

frodo-repo:


gandalf-repo (changes requested):

PR #1073, opened 2023-09-23
WEB-973 Feature: Introduce Localization Support

samwise-repo1:


gandalf-repo (changes requested):

PR #1144, opened 2023-09-23
WEB-1030 Feature: Introduce Mobile-First Responsive Design

samwise-repo1:


gandalf-repo (changes requested):

PR #1259, opened 2023-09-24
WEB-1127 Fix: Resolve Missing Icons in Navigation Bar

frodo-repo:


samwise-repo1 (approved):

PR #1290, opened 2023-09-24
WEB-1045 Feature: Introduce Mobile-First Responsive Design

frodo-repo:


gandalf-repo (commented):

PR #1314, opened 2023-09-24
WEB-1110 Feature: Introduce Mobile-First Responsive Design

samwise-repo1:


frodo-repo (approved):

PR #1342, opened 2023-09-24
WEB-1177 Style: Introduce New Color Palette for UI

samwise-repo1:


gandalf-repo (commented):

PR #1358, opened 2023-09-24
WEB-1147 Chore: Upgrade to Latest Node.js LTS Version

samwise-repo1:


frodo-repo (changes requested):
Consider renaming this variable to something more descriptive for better readability.

PR #1369, opened 2023-09-24
WEB-1162 Fix: Remove Redundant Conditional in Search Algorithm

frodo-repo:


gandalf-repo (changes requested):

PR #1432, opened 2023-09-25
WEB-1206 Fix: Resolve Missing Icons in Navigation Bar

gandalf-repo:


frodo-repo (changes requested):

PR #1438, opened 2023-09-25
WEB-1239 Fix: Resolve Missing Icons in Navigation Bar

frodo-repo:


samwise-repo1 (changes requested):

PR #1473, opened 2023-09-25
WEB-1238 Feature: Introduce Localization Support

samwise-repo1:


gandalf-repo (changes requested):

PR #1511, opened 2023-09-25
WEB-1276 Fix: Resolve Missing Icons in Navigation Bar

gandalf-repo:


frodo-repo (commented):

PR #1524, opened 2023-09-25
WEB-1255 Style: Introduce New Color Palette for UI

frodo-repo:


gandalf-repo (commented):

PR #1561, opened 2023-09-26
WEB-1314 Feature: Add Two-Factor Authentication

samwise-repo1:


gandalf-repo (changes requested):
Consider renaming this variable to something more descriptive for better readability.

PR #1665, opened 2023-09-26
WEB-1359 Refactor: Improve Code Readability in Helper Functions

frodo-repo:


gandalf-repo (changes requested):

PR #1724, opened 2023-09-26
WEB-1390 Fix: Resolve Missing Icons in Navigation Bar

frodo-repo:


gandalf-repo (commented):

PR #1725, opened 2023-09-26
WEB-1321 Bug: Fix Broken Image Links in Frontend

samwise-repo1:


gandalf-repo (changes requested):

PR #1813, opened 2023-09-27
WEB-1419 Chore: Update Project Dependencies

gandalf-repo:


frodo-repo (approved):

PR #1839, opened 2023-09-27
WEB-1417 Style: Introduce New Color Palette for UI

frodo-repo:


samwise-repo1 (changes requested):

PR #1858, opened 2023-09-27
WEB-1470 Style: Ensure Consistent Code Styling with Prettier

frodo-repo:


samwise-repo1 (changes requested):

PR #1863, opened 2023-09-27
WEB-1454 Test: Increase Test Coverage for API Routes

frodo-repo:
Mobile-first responsive design has been implemented to enhance user experience on mobile devices.


samwise-repo1 (changes requested):

PR #1865, opened 2023-09-27
WEB-1494 Chore: Update Project Dependencies

gandalf-repo:


samwise-repo1 (changes requested):

PR #1988, opened 2023-09-28
WEB-1588 Refactor: Split Monolithic Service into Microservices

frodo-repo:


samwise-repo1 (changes requested):

PR #2016, opened 2023-09-28
WEB-1614 Style: Introduce New Color Palette for UI

samwise-repo1:


gandalf-repo (changes requested):

PR #2044, opened 2023-09-29
WEB-1617 Feature: Introduce Localization Support

frodo-repo:


gandalf-repo (changes requested):

PR #2097, opened 2023-09-29
WEB-1675 Feature: Introduce Mobile-First Responsive Design

frodo-repo:


gandalf-repo (approved):

PR #2118, opened 2023-09-29
WEB-1670 Feature: Introduce Localization Support

frodo-repo:
This PR addresses a potential SQL injection vulnerability, thereby strengthening the security of our application.


samwise-repo1 (changes requested):

PR #2127, opened 2023-09-29
WEB-1670 Feature: Introduce Mobile-First Responsive Design

frodo-repo:


samwise-repo1 (approved):

PR #2157, opened 2023-09-29
WEB-1671 Style: Introduce New Color Palette for UI

samwise-repo1:


gandalf-repo (commented):

PR #2241, opened 2023-09-30
WEB-1754 Fix: Resolve Missing Icons in Navigation Bar

samwise-repo1:


gandalf-repo (changes requested):

PR #2291, opened 2023-09-30
WEB-1796 Fix: Resolve Missing Icons in Navigation Bar

frodo-repo:
All broken image links in the frontend have been fixed.


gandalf-repo (commented):

PR #2398, opened 2023-10-02
WEB-1890 Fix: Resolve Missing Icons in Navigation Bar

gandalf-repo:


frodo-repo (approved):

PR #2434, opened 2023-10-02
WEB-1935 Style: Introduce New Color Palette for UI

gandalf-repo:


samwise-repo1 (commented):

PR #2467, opened 2023-10-02
WEB-1937 Feature: Introduce Localization Support

samwise-repo1:


frodo-repo (changes requested):

PR #2514, opened 2023-10-03
WEB-1980 Refactor: Improve Code Readability in Helper Functions

gandalf-repo:


samwise-repo1 (approved):

PR #2527, opened 2023-10-03
WEB-2002 Fix: Resolve Missing Icons in Navigation Bar

samwise-repo1:


gandalf-repo (commented):

PR #2532, opened 2023-10-03
WEB-2010 Fix: Remove Redundant Conditional in Search Algorithm

gandalf-repo:


samwise-repo1 (approved):

PR #2572, opened 2023-10-03
WEB-2043 Style: Introduce New Color Palette for UI

samwise-repo1:


gandalf-repo (approved):

PR #2677, opened 2023-10-04
WEB-2089 Style: Introduce New Color Palette for UI

samwise-repo1:


samwise-repo1 (commented):
This new feature lacks unit tests. Could you add some to ensure everything works correctly?

PR #2717, opened 2023-10-05
WEB-2140 Style: Introduce New Color Palette for UI

gandalf-repo:


frodo-repo (changes requested):

PR #2722, opened 2023-10-05
WEB-2122 Fix: Remove Redundant Conditional in Search Algorithm

frodo-repo:


samwise-repo1 (changes requested):

PR #2814, opened 2023-10-06
WEB-2188 Style: Introduce New Color Palette for UI

samwise-repo1:


frodo-repo (changes requested):

PR #2825, opened 2023-10-07
WEB-2200 Refactor: Split Monolithic Service into Microservices

gandalf-repo:


frodo-repo (approved):

PR #2891, opened 2023-10-09
WEB-2262 Style: Introduce New Color Palette for UI

samwise-repo1:


gandalf-repo (commented):

PR #2906, opened 2023-10-10
WEB-2269 Feature: Introduce Localization Support

samwise-repo1:


gandalf-repo (approved):

PR #2910, opened 2023-10-10
WEB-2278 Fix: Resolve Missing Icons in Navigation Bar

samwise-repo1:


gandalf-repo (changes requested):

PR #2954, opened 2023-10-15
WEB-2320 Style: Introduce New Color Palette for UI

samwise-repo1:


frodo-repo (changes requested):

PR #3040, opened 2023-11-06

frodo-repo:


samwise-repo1 (changes requested):
SCRIPT-This function seems to have grown quite complex. Can we break it down further?

PR #3042, opened 2023-11-07
SCRIPT-WEB-2355 Fix: Resolve Missing Icons in Navigation Bar

gandalf-repo:


samwise-repo1 (changes requested):
SCRIPT-I'm a bit concerned about the potential performance impact of this loop. Have you thought about optimizing it?

PR #3057, opened 2023-11-10
SCRIPT-WEB-2418 Chore: Upgrade to Latest Node.js LTS Version

gandalf-repo:


gandalf.js
```diff
@@ -1,121 +1,128 @@
 /**
```
frodo-repo, line 1: This block of code could benefit from some comments explaining the logic behind it.

PR #3058, opened 2023-11-10

gandalf-repo:


gandalf.js
```diff
@@ -1,104 +1,105 @@
 /**
```
frodo-repo, line 1: This new feature lacks unit tests. Could you add some to ensure everything works correctly?

PR #3060, opened 2023-11-11
SCRIPT-WEB-2443 Bug: Resolve Inconsistent User Session Handling

samwise-repo1:


samwise.js
```diff
@@ -5,17 +5,20 @@
  * Released under the MIT license.
```
gandalf-repo, line 5: Good job on updating the documentation along with the changes. It's very thorough and helpful!

PR #3063, opened 2023-11-12
SCRIPT-WEB-2423 Feature: Implement User Activity Dashboard

gandalf-repo:


gandalf.js
```diff
@@ -1,51 +1,45 @@
 /**
```
samwise-repo1, line 1: Consider extracting this repeated logic into a separate function to avoid duplication.

PR #3070, opened 2023-11-14
SCRIPT-WEB-2444 Style: Ensure Consistent Code Styling with Prettier

frodo-repo:


frodo.js
```diff
@@ -1,15 +1,16 @@
 /**
```
samwise-repo1, line 1: This hardcoded value might be better suited as a constant at the top of the file.

PR #3071, opened 2023-11-14
Bug: Fix Broken Image Links in Frontend

samwise-repo1:


samwise.js
```diff
@@ -1,26 +1,24 @@
 /**
```
frodo-repo, line 1: SCRIPT-I like how you've used promise chaining here, it makes the code flow very clear.

PR #3072, opened 2023-11-14
SCRIPT-WEB-2442 Refactor: Split Monolithic Service into Microservices

gandalf-repo:


gandalf.js
```diff
@@ -2,132 +2,125 @@
  * Extras Plugin
```
gandalf-repo, line 2: This function seems to have grown quite complex. Can we break it down further?

PR #3078, opened 2023-11-16
SCRIPT-WEB-2453-SCRIPT-Bug: Fix Error in Date Formatting in Reports

samwise-repo1:


samwise.js
```diff
@@ -1,13 +1,13 @@
 /**
```
frodo-repo, line 1: SCRIPT-I'm a bit concerned about the potential performance impact of this loop. Have you thought about optimizing it?

PR #3086, opened 2023-11-18
SCRIPT-WEB-2437-SCRIPT-Bug: Fix Error in Date Formatting in Reports

gandalf-repo:


gandalf.js
```diff
@@ -1,43 +1,45 @@
-const filteredArray = array.filter(item => item !== null && item !== undefined);
```

PR #3089, opened 2023-11-19
SCRIPT-WEB-2430-SCRIPT-Fix: Resolve Broken Pagination in Blog Section

frodo-repo:

### DIFF
--- a/frodo.js
+++ b/frodo.js
@@ -1,133 +1,131 @@
 /**
  * Extras Plugin
  *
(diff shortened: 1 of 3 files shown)


frodo.js
```diff
@@ -1,133 +1,131 @@
 /**
```
samwise-repo1, line 1: SCRIPT-This function seems to have grown quite complex. Can we break it down further?

PR #3095, opened 2023-11-21
SCRIPT-WEB-2469-SCRIPT-Bug: Resolve Inconsistent User Session Handling

samwise-repo1:


samwise.js
```diff
@@ -6,30 +6,32 @@
  */
```
frodo-repo, line 6: SCRIPT-Consider renaming this variable to something more descriptive for better readability.

PR #3161, opened 2023-12-13
SCRIPT-WEB-2549-SCRIPT-Chore: Add Linting to CI Pipeline

samwise-repo1:


samwise.js
```diff
@@ -3,32 +3,32 @@
  *
```
gandalf-repo, line 3: SCRIPT-This logic might be clearer if you used a switch statement instead of multiple if-else statements.

PR #3234, opened 2024-01-18
SCRIPT-WEB-2532-SCRIPT-Test: Add Unit Tests for New String Manipulation Functions

gandalf-repo:


gandalf.js
```diff
@@ -1,134 +1,124 @@
-const isString = typeof variable === "string";
```
gandalf-repo, line 1: SCRIPT-Really good job with this regex. It covers all the cases we discussed.

PR #3244, opened 2024-01-21
SCRIPT-WEB-2629-SCRIPT-Refactor: Switch from Callbacks to Promises

frodo-repo:

### DIFF
--- a/frodo.js
+++ b/frodo.js
@@ -1,101 +1,106 @@
 /**
  * Extras Plugin
- *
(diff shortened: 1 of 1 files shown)


frodo.js
```diff
@@ -1,101 +1,106 @@
 /**
```
frodo-repo, line 1: SCRIPT-This hardcoded value might be better suited as a constant at the top of the file.

PR #3252, opened 2024-01-26
SCRIPT-WEB-2626-SCRIPT-Refactor: Switch from Callbacks to Promises

frodo-repo:


frodo.js
```diff
@@ -13,123 +13,112 @@
```
samwise-repo1, line 13: SCRIPT-Your solution to the bug is very elegant, well done!

PR #3256, opened 2024-01-29
SCRIPT-WEB-2639-SCRIPT-Chore: Update Project Dependencies

samwise-repo1:


samwise.js
```diff
@@ -10,99 +10,104 @@
     "use strict";
```
gandalf-repo, line 10: SCRIPT-This function seems to have grown quite complex. Can we break it down further?

PR #3261, opened 2024-02-02
SCRIPT-WEB-2590-SCRIPT-Bug: Fix Error in Date Formatting in Reports

samwise-repo1:


samwise.js
```diff
@@ -6,106 +6,100 @@
  */
```
samwise-repo1, line 6: SCRIPT-This function might benefit from using async/await instead of callbacks for better readability.

PR #3263, opened 2024-02-03
SCRIPT-WEB-2669-SCRIPT-Refactor: Split Monolithic Service into Microservices

frodo-repo:
SCRIPT-This PR resolves an issue where icons in the navigation bar were missing due to incorrect paths.


frodo.js
```diff
@@ -1,39 +1,37 @@
-/**
```
frodo-repo, line 1: SCRIPT-This function seems to have grown quite complex. Can we break it down further?

PR #3282, opened 2024-02-07
SCRIPT-WEB-2567-SCRIPT-Chore: Add Linting to CI Pipeline

samwise-repo1:


samwise.js
```diff
@@ -1,50 +1,51 @@
-/**
```

PR #3289, opened 2024-02-10
SCRIPT-WEB-2683-SCRIPT-Chore: Update Project Dependencies

frodo-repo:


frodo.js
```diff
@@ -2,111 +2,112 @@
  * Extras Plugin
```
samwise-repo1, line 2: SCRIPT-This function seems to have grown quite complex. Can we break it down further?

PR #3355, opened 2024-03-08
SCRIPT-WEB-2722-SCRIPT-Docs: Update API Documentation for Recent Changes

frodo-repo:

### DIFF
--- a/frodo.js
+++ b/frodo.js
@@ -6,30 +6,26 @@
  *
  * See README.md for details.
  *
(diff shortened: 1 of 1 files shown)


frodo.js
```diff
@@ -6,40 +6,34 @@
  *
```
samwise-repo1, line 6: SCRIPT-Consider extracting this repeated logic into a separate function to avoid duplication.

PR #3374, opened 2024-03-11
SCRIPT-WEB-2761-SCRIPT-Bug: Fix Broken Image Links in Frontend

frodo-repo:


frodo.js
```diff
@@ -1,28 +1,30 @@
-/**
```
gandalf-repo, line 1: SCRIPT-This hardcoded value might be better suited as a constant at the top of the file.

PR #3381, opened 2024-03-13
SCRIPT-Chore: Add Linting to CI Pipeline

gandalf-repo:


gandalf.js
```diff
@@ -1,119 +1,112 @@
 /**
```
gandalf-repo, line 1: SCRIPT-This line seems to be longer than our standard line length. Could you split it up for better readability?

PR #3421, opened 2024-03-20
SCRIPT-Fix: Resolve Broken Pagination in Blog Section

gandalf-repo:


gandalf.js
```diff
@@ -1,8 +1,8 @@
 /**
```

PR #3425, opened 2024-03-20
SCRIPT-WEB-2799-SCRIPT-Refactor: Switch from Callbacks to Promises

samwise-repo1:


samwise.js
```diff
@@ -1,5 +1,5 @@
 /**
```
samwise-repo1, line 1: SCRIPT-Good job on updating the documentation along with the changes. It's very thorough and helpful!

PR #3432, opened 2024-03-22
SCRIPT-WEB-2676-SCRIPT-Bug: Fix Broken Image Links in Frontend

samwise-repo1:


samwise.js
```diff
@@ -5,7 +5,8 @@
  * Released under the MIT license.
```
samwise-repo1, line 5: SCRIPT-I'm a bit concerned about the potential performance impact of this loop. Have you thought about optimizing it?

PR #3439, opened 2024-03-23
SCRIPT-WEB-2836-SCRIPT-Feature: Add Two-Factor Authentication

gandalf-repo:


gandalf.js
```diff
@@ -1,21 +1,20 @@
 /**
```
gandalf-repo, line 1: SCRIPT-This line seems to be longer than our standard line length. Could you split it up for better readability?

PR #3441, opened 2024-03-24
SCRIPT-WEB-2810-SCRIPT-Fix: Resolve Missing Icons in Navigation Bar

samwise-repo1:


samwise.js
```diff
@@ -1,75 +1,74 @@
 /**
```
frodo-repo, line 1: SCRIPT-This logic might be clearer if you used a switch statement instead of multiple if-else statements.

PR #3446, opened 2024-03-25
SCRIPT-WEB-2838-SCRIPT-Refactor: Improve Code Readability in Helper Functions

gandalf-repo:


gandalf.js
```diff
@@ -1,7 +1,7 @@
 /**
```

PR #3494, opened 2024-04-08
SCRIPT-WEB-2870-SCRIPT-Refactor: Split Monolithic Service into Microservices

gandalf-repo:


gandalf.js
```diff
@@ -1,5 +1,5 @@
 /**
```
frodo-repo, line 1: SCRIPT-This function seems to have grown quite complex. Can we break it down further?

PR #3509, opened 2024-04-10
SCRIPT-WEB-2854-SCRIPT-Refactor: Switch from Callbacks to Promises

gandalf-repo:

### DIFF
--- a/gandalf.js
+++ b/gandalf.js
@@ -1,130 +1,130 @@
 /**
  * Extras Plugin
  *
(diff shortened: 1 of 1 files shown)


gandalf.js
```diff
@@ -1,130 +1,130 @@
 /**
```
samwise-repo1, line 1: SCRIPT-I like how you've used promise chaining here, it makes the code flow very clear.

PR #3534, opened 2024-04-14
SCRIPT-WEB-2926-SCRIPT-Style: Ensure Consistent Code Styling with Prettier

gandalf-repo:


gandalf.js
```diff
@@ -11,19 +11,22 @@
  */
```
frodo-repo, line 11: SCRIPT-This function might benefit from using async/await instead of callbacks for better readability.

PR #3546, opened 2024-04-16
SCRIPT-WEB-2957-SCRIPT-Chore: Update Project Dependencies

frodo-repo:

### DIFF
--- a/frodo.js
+++ b/frodo.js
@@ -1,20 +1,19 @@
 /**
  * Extras Plugin
  *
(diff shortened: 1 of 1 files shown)


frodo.js
```diff
@@ -1,20 +1,19 @@
 /**
```
samwise-repo1, line 1: SCRIPT-Your solution to the bug is very elegant, well done!

PR #3554, opened 2024-04-17
SCRIPT-WEB-2937-SCRIPT-Test: Add Unit Tests for New String Manipulation Functions

samwise-repo1:


samwise.js
```diff
@@ -1,9 +1,8 @@
 /**
```
frodo-repo, line 1: SCRIPT-I like how you've used promise chaining here, it makes the code flow very clear.

PR #3573, opened 2024-04-20
SCRIPT-WEB-2992-SCRIPT-Refactor: Switch from Callbacks to Promises

frodo-repo:


frodo.js
```diff
@@ -3,82 +3,88 @@
  *
```
samwise-repo1, line 3: SCRIPT-This function might benefit from using async/await instead of callbacks for better readability.

PR #3589, opened 2024-04-25
SCRIPT-WEB-2971-SCRIPT-Style: Introduce New Color Palette for UI

samwise-repo1:


samwise.js
```diff
@@ -5,30 +5,29 @@
  * Released under the MIT license.
```
samwise-repo1, line 5: SCRIPT-Excellent use of destructuring here, it makes the code much cleaner.

PR #3595, opened 2024-04-29
SCRIPT-Performance: Improve Efficiency of Sorting Algorithm

gandalf-repo:


gandalf.js
```diff
@@ -1,133 +1,127 @@
 /**
```

PR #3597, opened 2024-04-30
SCRIPT-WEB-2993-SCRIPT-Bug: Resolve Inconsistent User Session Handling

frodo-repo:


frodo.js
```diff
@@ -1,55 +1,55 @@
 /**
```
gandalf-repo, line 1: SCRIPT-This logic might be clearer if you used a switch statement instead of multiple if-else statements.

PR #3616, opened 2024-05-04
SCRIPT-WEB-2991-SCRIPT-Bug: Resolve Inconsistent User Session Handling

samwise-repo1:


samwise.js
```diff
@@ -1,60 +1,59 @@
 /**
```
gandalf-repo, line 1: SCRIPT-Your solution to the bug is very elegant, well done!

PR #3619, opened 2024-05-05
SCRIPT-WEB-3037-SCRIPT-Bug: Fix Broken Image Links in Frontend

samwise-repo1:

### DIFF
--- a/samwise.js
+++ b/samwise.js
@@ -9,11 +9,11 @@
 ( function( document ) {
     "use strict";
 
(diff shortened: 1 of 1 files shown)


samwise.js
```diff
@@ -9,44 +9,49 @@
 ( function( document ) {
```
gandalf-repo, line 9: SCRIPT-I'm a bit concerned about the potential performance impact of this loop. Have you thought about optimizing it?

PR #3638, opened 2024-05-11
SCRIPT-WEB-3062-SCRIPT-Bug: Resolve Inconsistent User Session Handling

samwise-repo1:


samwise.js
```diff
@@ -3,180 +3,168 @@
  *
```
samwise-repo1, line 3: SCRIPT-Good job on updating the documentation along with the changes. It's very thorough and helpful!

PR #3664, opened 2024-05-23
SCRIPT-WEB-3101-SCRIPT-Chore: Upgrade to Latest Node.js LTS Version

gandalf-repo:


gandalf.js
```diff
@@ -3,8 +3,8 @@
  *
```
samwise-repo1, line 3: SCRIPT-I'm a bit concerned about the potential performance impact of this loop. Have you thought about optimizing it?

PR #3665, opened 2024-05-23
SCRIPT-WEB-2896-SCRIPT-Bug: Fix Broken Image Links in Frontend

gandalf-repo:


gandalf.js
```diff
@@ -3,122 +3,131 @@
  *
```
samwise-repo1, line 3: SCRIPT-This logic might be clearer if you used a switch statement instead of multiple if-else statements.

PR #3678, opened 2024-05-26
SCRIPT-WEB-3110-SCRIPT-Feature: Introduce Localization Support

gandalf-repo:


gandalf.js
```diff
@@ -1,18 +1,20 @@
-const largestNumber = Math.max(...numbers);
```

PR #3679, opened 2024-05-26
SCRIPT-WEB-3057-SCRIPT-Chore: Add Linting to CI Pipeline

samwise-repo1:


samwise.js
```diff
@@ -1,23 +1,22 @@
 /**
```
frodo-repo, line 1: SCRIPT-This line seems to be longer than our standard line length. Could you split it up for better readability?

PR #3694, opened 2024-05-29
SCRIPT-WEB-3085-SCRIPT-Style: Ensure Consistent Code Styling with Prettier

gandalf-repo:


gandalf.js
```diff
@@ -1,29 +1,30 @@
-/**
```
samwise-repo1, line 1: SCRIPT-This hardcoded value might be better suited as a constant at the top of the file.

PR #3700, opened 2024-05-30
SCRIPT-WEB-3089-SCRIPT-Performance: Improve Efficiency of Sorting Algorithm

samwise-repo1:


samwise.js
```diff
@@ -2,14 +2,14 @@
  * Autoplay plugin - Automatically advance slideshow after N seconds
```
frodo-repo, line 2: SCRIPT-This hardcoded value might be better suited as a constant at the top of the file.

PR #3715, opened 2024-06-04
SCRIPT-WEB-3074-SCRIPT-Bug: Fix Broken Image Links in Frontend

samwise-repo1:


samwise.js
```diff
@@ -1,56 +1,55 @@
 /**
```
samwise-repo1, line 1: SCRIPT-This function might benefit from using async/await instead of callbacks for better readability.

PR #3732, opened 2024-06-15
SCRIPT-WEB-3029-SCRIPT-Test: Add Unit Tests for New String Manipulation Functions

gandalf-repo:


gandalf.js
```diff
@@ -8,126 +8,125 @@
  *
```
samwise-repo1, line 8: SCRIPT-This line seems to be longer than our standard line length. Could you split it up for better readability?

PR #3747, opened 2024-06-21
SCRIPT-WEB-3162-SCRIPT-Bug: Resolve Inconsistent User Session Handling

samwise-repo1:
SCRIPT-This PR fixes an error in date formatting in our reports, improving data accuracy.


samwise.js
```diff
@@ -1,17 +1,17 @@
 /**
```

PR #3754, opened 2024-06-24
SCRIPT-WEB-3172-SCRIPT-Feature: Add Export Data Functionality

samwise-repo1:


samwise.js
```diff
@@ -1,19 +1,18 @@
 /**
```
frodo-repo, line 1: SCRIPT-This logic might be clearer if you used a switch statement instead of multiple if-else statements.

PR #3785, opened 2024-06-28
SCRIPT-WEB-3115-SCRIPT-Bug: Resolve Inconsistent User Session Handling

gandalf-repo:


gandalf.js
```diff
@@ -4,97 +4,104 @@
  * This plugin performs initialization (like calling mermaid.initialize())
```
frodo-repo, line 4: SCRIPT-This function might benefit from using async/await instead of callbacks for better readability.

PR #3786, opened 2024-06-28
SCRIPT-WEB-3071-SCRIPT-Bug: Resolve Inconsistent User Session Handling

samwise-repo1:


samwise.js
```diff
@@ -1,38 +1,39 @@
 /**
```
samwise-repo1, line 1: SCRIPT-This hardcoded value might be better suited as a constant at the top of the file.

PR #3794, opened 2024-06-30
SCRIPT-WEB-3034-SCRIPT-Bug: Fix Broken Image Links in Frontend

samwise-repo1:


samwise.js
```diff
@@ -1,15 +1,14 @@
 /**
```
gandalf-repo, line 1: SCRIPT-I like how you've used promise chaining here, it makes the code flow very clear.

PR #3832, opened 2024-07-07
SCRIPT-WEB-3155-SCRIPT-Chore: Add Linting to CI Pipeline

samwise-repo1:


samwise.js
```diff
@@ -4,25 +4,23 @@
  * Copyright 2016 Henrik Ingo, henrik.ingo@avoinelama.fi
```
samwise-repo1, line 4: SCRIPT-Your solution to the bug is very elegant, well done!

PR #3855, opened 2024-07-13
SCRIPT-WEB-3178-SCRIPT-Bug: Fix Broken Image Links in Frontend

samwise-repo1:


samwise.js
```diff
@@ -1,5 +1,6 @@
-/**
```

PR #3862, opened 2024-07-15
SCRIPT-WEB-3266-SCRIPT-Bug: Resolve Inconsistent User Session Handling

gandalf-repo:


gandalf.js
```diff
@@ -1,109 +1,99 @@
 /**
```
gandalf-repo, line 1: SCRIPT-This function seems to have grown quite complex. Can we break it down further?

PR #3881, opened 2024-07-28
SCRIPT-WEB-3235-SCRIPT-Chore: Add Linting to CI Pipeline

gandalf-repo:

### DIFF
--- a/gandalf.js
+++ b/gandalf.js
@@ -6,59 +6,64 @@
  *
  * See README.md for details.
  *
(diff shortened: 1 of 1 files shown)


gandalf.js
```diff
@@ -6,59 +6,64 @@
  *
```
gandalf-repo, line 6: SCRIPT-Your use of CSS grid here makes the layout very flexible. Good job!

PR #3928, opened 2024-08-10
SCRIPT-WEB-3336-SCRIPT-Bug: Fix Broken Image Links in Frontend

samwise-repo1:


samwise.js
```diff
@@ -1,4 +1,4 @@
-/**
```
samwise-repo1, line 1: SCRIPT-This new dependency seems to be unused. Can you please confirm?

PR #3954, opened 2024-08-16
SCRIPT-WEB-3383-SCRIPT-Feature: Add Export Data Functionality

gandalf-repo:

### DIFF
--- a/gandalf.js
+++ b/gandalf.js
@@ -6,18 +6,17 @@
  *
  * See README.md for details.
  *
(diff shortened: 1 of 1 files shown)


gandalf.js
```diff
@@ -6,101 +6,104 @@
  *
```
gandalf-repo, line 6: SCRIPT-This logic might be clearer if you used a switch statement instead of multiple if-else statements.

PR #3955, opened 2024-08-16
SCRIPT-WEB-3378-SCRIPT-Chore: Upgrade to Latest Node.js LTS Version

samwise-repo1:


samwise.js
```diff
@@ -5,37 +5,36 @@
  * Released under the MIT license.
```
frodo-repo, line 5: SCRIPT-This hardcoded value might be better suited as a constant at the top of the file.

PR #3958, opened 2024-08-17
SCRIPT-WEB-3386-SCRIPT-Refactor: Switch from Callbacks to Promises

samwise-repo1:


samwise.js
```diff
@@ -2,20 +2,21 @@
  * Autoplay plugin - Automatically advance slideshow after N seconds
```
samwise-repo1, line 2: SCRIPT-Your use of CSS grid here makes the layout very flexible. Good job!

PR #3960, opened 2024-08-17
SCRIPT-WEB-3326-SCRIPT-Refactor: Switch from Callbacks to Promises

gandalf-repo:


gandalf.js
```diff
@@ -2,108 +2,117 @@
  * Extras Plugin
```
samwise-repo1, line 2: SCRIPT-Really good job with this regex. It covers all the cases we discussed.

PR #3961, opened 2024-08-17
SCRIPT-WEB-3361-SCRIPT-Fix: Resolve Missing Icons in Navigation Bar

frodo-repo:


frodo.js
```diff
@@ -1,126 +1,130 @@
-/**
```
frodo-repo, line 1: SCRIPT-This hardcoded value might be better suited as a constant at the top of the file.

PR #3971, opened 2024-08-20
SCRIPT-Feature: Introduce Localization Support

gandalf-repo:


gandalf.js
```diff
@@ -1,129 +1,128 @@
 /**
```
gandalf-repo, line 1: SCRIPT-Your solution to the bug is very elegant, well done!

PR #3990, opened 2024-08-25
SCRIPT-WEB-3199-SCRIPT-Feature: Add Two-Factor Authentication

gandalf-repo:


gandalf.js
```diff
@@ -5,96 +5,101 @@
  * for the extras/ plugins if they are loaded into a presentation.
```
frodo-repo, line 5: SCRIPT-Excellent use of destructuring here, it makes the code much cleaner.

PR #4474, opened 2025-01-25
SCRIPT-Bug: Fix Error in Date Formatting in Reports

gandalf-repo:


gandalf.js
```diff
@@ -1,25 +1,25 @@
 /**
```
frodo-repo, line 1: SCRIPT-Really good job with this regex. It covers all the cases we discussed.

PR #4496, opened 2025-01-30
SCRIPT-WEB-3857-SCRIPT-Feature: Add Export Data Functionality

gandalf-repo:


gandalf.js
```diff
@@ -2,56 +2,59 @@
  * Extras Plugin
```
gandalf-repo, line 2: SCRIPT-This logic might be clearer if you used a switch statement instead of multiple if-else statements.

PR #4504, opened 2025-02-03
SCRIPT-WEB-3902-SCRIPT-Docs: Update API Documentation for Recent Changes

gandalf-repo:


gandalf.js
```diff
@@ -1,17 +1,17 @@
 /**
```

PR #4536, opened 2025-02-17
SCRIPT-WEB-3879-SCRIPT-Style: Introduce New Color Palette for UI

gandalf-repo:


gandalf.js
```diff
@@ -9,52 +9,54 @@
  * Copyright 2016 Henrik Ingo (@henrikingo)
```
gandalf-repo, line 9: SCRIPT-Really good job with this regex. It covers all the cases we discussed.

PR #4576, opened 2025-02-28
SCRIPT-WEB-4014-SCRIPT-Chore: Move Project Hosting from Heroku to AWS

gandalf-repo:


gandalf.js
```diff
@@ -1,25 +1,26 @@
 /**
```
gandalf-repo, line 1: SCRIPT-This function seems to have grown quite complex. Can we break it down further?

PR #4724, opened 2025-06-05
SCRIPT-WEB-3940-SCRIPT-Docs: Update API Documentation for Recent Changes

samwise-repo1:


samwise.js
```diff
@@ -1,5 +1,5 @@
 /**
```
gandalf-repo, line 1: SCRIPT-I like how you've used promise chaining here, it makes the code flow very clear.

PR #4760, opened 2025-06-12
SCRIPT-WEB-3980-SCRIPT-Feature: Introduce Localization Support

samwise-repo1:


samwise.js
```diff
@@ -6,29 +6,29 @@
  */
```
frodo-repo, line 6: SCRIPT-Your solution to the bug is very elegant, well done!

PR #4776, opened 2025-06-15
SCRIPT-WEB-4224-SCRIPT-Chore: Add Linting to CI Pipeline

samwise-repo1:


samwise.js
```diff
@@ -3,62 +3,58 @@
  *
```
samwise-repo1, line 3: SCRIPT-This line seems to be longer than our standard line length. Could you split it up for better readability?

PR #4823, opened 2025-06-27
SCRIPT-WEB-4265-SCRIPT-Fix: Resolve Missing Icons in Navigation Bar

frodo-repo:


frodo.js
```diff
@@ -2,128 +2,130 @@
  * Extras Plugin
```
gandalf-repo, line 2: SCRIPT-This function seems to have grown quite complex. Can we break it down further?

PR #4944, opened 2025-07-31
SCRIPT-WEB-4298-SCRIPT-Bug: Fix Broken Image Links in Frontend

gandalf-repo:

### DIFF
--- a/gandalf.js
+++ b/gandalf.js
@@ -1,66 +1,62 @@
 /**
  * Extras Plugin
  *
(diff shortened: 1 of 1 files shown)


gandalf.js
```diff
@@ -1,78 +1,73 @@
 /**
```
gandalf-repo, line 1: SCRIPT-This hardcoded value might be better suited as a constant at the top of the file.

PR #5027, opened 2025-08-27
SCRIPT-WEB-4417-SCRIPT-Feature: Introduce Mobile-First Responsive Design

samwise-repo1:


samwise.js
```diff
@@ -1,31 +1,30 @@
-/**
```
frodo-repo, line 1: SCRIPT-This hardcoded value might be better suited as a constant at the top of the file.

